PR #8, opened 2019-08-29
Completed DFS-2 : Please review

muditabysani:


muditabysani:
@sehaj-grover Please review the DFS approach. There is some issue with it and I'm not able to figure it out

Problem1.py
```diff
@@ -0,0 +1,61 @@
...
+			for j in range(columns):
+				if grid[i][j] == "1":
+					res += 1
+					self.dfs(grid, i, j)
+		return res
+	
+	def dfs(self, grid, x, y):
+		if x < 0 or y < 0 or x >= len(grid) or y >= len(grid[0]) or grid[x][y] != "1":
+			return
+		grid[x][y] == "0"
```
sehaj-grover: it should be grid[x][y]='0'

muditabysani: In python, single quote and double quote doesn't have any difference

muditabysani: I'm having issue with the if condition and that is the maximum recursion depth exceeded in comparison

sehaj-grover: what I meant was instead of changing grid[x][y] to 0 you are actually comparing its value to =='0'

muditabysani: I'm sorry... my bad.

PR #15, opened 2019-09-20
add number of islands

pradhan1234:


pradhan1234:
this PR was closed, added as per requirements here #16 
thanks

PR #18, opened 2019-09-20
Problem-1 solution added

akhileshborgaonkar:
Problem-1 DFS solution. LC verified.

### DIFF
--- a/FindIslands.java
+++ b/FindIslands.java
@@ -1,0 +1,51 @@
+/*
+Author: Akhilesh Borgaonkar
+Problem: Given a 2d grid map of '1's (land) and '0's (water), count the number of islands.
+Approach: I have used DFS approach here. I am exploring the input grid to find '1' and then calling dfs function to explore its neighbors.
+    While visiting the neighbors I am marking those cells as '0' to avoid revisiting them again. While in dfs function, I check for edge cases of grid coordinates
+    and if the cell has '1' char then mark it as '0'.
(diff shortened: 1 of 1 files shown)


akhileshborgaonkar:
Problem-2 String decoding has a known issue. In the case of nested square brackets, output string returns in reverse order. Else, works fine.

DecodeString.java
```diff
@@ -0,0 +1,44 @@
...
+
+        Stack<Integer> frequency = new Stack<Integer>();
+        Stack<String> strings = new Stack<String>();
+        String currStr;
+        String output = "";
+        int currFreq=0;
+
+        for(int i=0; i<s.length() ; i++){
+            if(Character.isDigit(s.charAt(i))){
+                currFreq = Character.getNumericValue(s.charAt(i));
```
sehaj-grover: no correct for 2 digit numbers 
for 13 your number is only 1

DecodeString.java
```diff
@@ -0,0 +1,44 @@
...
+                System.out.println(currFreq);
+            }
+            else if(s.charAt(i)=='[') {
+                frequency.push(currFreq);
+
+                currStr="";
+                currFreq=0;
+            }
+            else if(s.charAt(i)==']') {
+                currStr = strings.pop();
```
sehaj-grover: only taking one char value for [bc] you are taking only c

akhileshborgaonkar:
> So can you fix the algorithm now?

Yes, it is fixed now.

sehaj-grover:
Great!

PR #48, opened 2019-12-31
completed DFS-2

AvinavTyagi:


NumIslands.java
```diff
@@ -0,0 +1,89 @@
+/*Time Complexity: O(m*n)
+Space Complexity: O(m*n)
+Accepted on Leetcode: Yes
+Approach: DFS Solution: Imagining a graph instead of a 2D array. A node containing a '1' is a root node that triggers a Depth First Search.
```
neha79091: space complexity should be constant

AvinavTyagi: Yes, you're right. I'll correct it

PR #52, opened 2020-01-25
dfs-2 problems num of islands and decode string

vyomgupta1:
Kindly review.

### DIFF
--- a/problem-1_islands.java
+++ b/problem-1_islands.java
@@ -1,0 +1,43 @@
+// Time Complexity :
+//      m - number of rows
+//      n - number of columns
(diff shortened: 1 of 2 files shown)


problem-1_islands.java
```diff
@@ -0,0 +1,43 @@
+// Time Complexity :
+//      m - number of rows
+//      n - number of columns
+//      total time complexity will be O(m*n)
+//
+// Space Complexity :
+//      O(m+n)->n is number of columns. m is rows. Asymptotically it is linear space.
+//
```
neha79091: In case of a recursive solution,it is m*n

PR #61, opened 2020-01-29
DFS-2 Solutions

kamaleshanantha:


numberOfIslands.py
```diff
@@ -0,0 +1,31 @@
+"""
+Time complexity: O(M * N)
+Space complexity: O( sqrt(M^2 + N^2))
+Compiled on leetcode: Yes
+Difficulties faced: No
+"""
+from collections import deque
```
neha79091: Try solving this using DFS approach also

numberOfIslands.py
```diff
@@ -0,0 +1,31 @@
+"""
+Time complexity: O(M * N)
+Space complexity: O( sqrt(M^2 + N^2))
+Compiled on leetcode: Yes
+Difficulties faced: No
```
neha79091: Try to do it with DFS also

PR #71, opened 2020-03-06
Completed problem 1

PratikshaPP:
Completed problem 1. Couldn't completely solve problem 2 . Need assistance

### DIFF
--- a/decodestrings.py
+++ b/decodestrings.py
@@ -1,0 +1,38 @@
+// Time Complexity : Solution incomplete
+// Space Complexity : Solution incomplete
+// Did this code successfully run on Leetcode : No
+// Any problem you faced while coding this : Yes, couldn't completely solve the question
+
+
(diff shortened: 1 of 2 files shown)


decodestrings.py
```diff
@@ -0,0 +1,38 @@
...
+        
+        if s[index].isdigit():
+            num = num*10 + int(s[index])
+            
+        if s[index].isalpha():
+            intermediate += s[index]
+            
+        if s[index] == '[':
+            num = 0
+            intermediate = ""
```
Lohith-Aswathappa: when you encounter the '[' you set it empty string and num = 0 without even processing 
then when it reaches the ']' you should return until all the elements are processed in the brackets but then the condition terms only when index reaches the full length of the string. Let me know if you need more help.

Lohith-Aswathappa:
Good work. I have added inline comments for the 1st problem. Do check

PR #74, opened 2020-03-06
Completed number of islands

arnav1996:


Lohith-Aswathappa:
Good work Arnav!
Submit the decode strings problem as well

PR #85, opened 2020-04-17
adding Q1

RutujaKaushike:


devashishnyati:
Good job Rutuja!
Nice work with the BFS solution.

PR #89, opened 2020-04-18
Added Solution to Problem 1

kannan-mrugapphan:
passes all test cases on leetcode


devashishnyati:
Good job Kannan!

kannan-mrugapphan:
Added solution to problem 2 - please review

PR #215, opened 2020-10-17
DFS-2 : Please review

neharikakhera:


Decode string
```diff
@@ -0,0 +1,47 @@
...
+            }
+            else if(s.charAt(i) == ']') {
+                int count = countStack.pop();
+                String temp = tempStack.pop();
+                for(int k = 0; k < count; k++){
+                    output.append(temp);
+                }
+             }
+            
+            i++;
```
Lohith-Aswathappa, line 43: this is an extra increment, you need to avoid this.

Decode string
```diff
@@ -0,0 +1,47 @@
...
+                 tempStack.push(temp);
+                }
+               
+            }
+            else if(s.charAt(i) == ']') {
+                int count = countStack.pop();
+                String temp = tempStack.pop();
+                for(int k = 0; k < count; k++){
+                    output.append(temp);
+                }
```
Lohith-Aswathappa, line 40: after this for operation, i still at the same location. 
Since you have popped the stack, popping again will give you an empty stack error. 

The logic only for normal cases. Nested brackets need to be taken care of and retry the problem again

Lohith-Aswathappa (commented):
Nice try, Neharika!
Review the inline comment and try the problems again

Lohith-Aswathappa:
Nice try, Neharika!
Review the inline comment and try the problems again

PR #235, opened 2020-11-16
DFS-2

sumitjanawlekar:
Not submitting the recurssive solution for decode-string problem, as I am having difficulties understanding it.
Rest solutions are there in this PR


BFS_number_of_islands.java
```diff
@@ -0,0 +1,58 @@
+// Time Complexity : O(m*n), where m is the number of rows and n is the number of columns in the grid (in the worst case O(2m*n))
+// Space Complexity :O(m), where m is the minimum of the rows and colmns in the grid (space for the queue)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : I wanted to know is it ok to use the same letter with different meanings in the time and
```
kannan-mrugapphan, line 4: Its better to stick with same symbols in both time and space costs

kannan-mrugapphan (commented):
Good Job

kannan-mrugapphan:
Good Job

PR #263, opened 2020-12-25
DFS 2

rishiraj824:


ramya-sruthi-d (commented):
Good, the tc for num of islands problem is O(m*n)

ramya-sruthi-d:
Good, the tc for num of islands problem is O(m*n)

PR #343, opened 2021-03-23
Completed DFS-2

Tirthaj:


Number of Islands.py
```diff
@@ -0,0 +1,24 @@
+'''
+TC: O(n)
+SC: O(n)
+'''
```
Lohith-Aswathappa, line 4: mention what your N stands for 

TIme complexity would be O(MxN)
and space (1) since you are not using any extra space.

Lohith-Aswathappa (commented):
Good work, Tirthaj!
Check inline comments

Lohith-Aswathappa:
Good work, Tirthaj!
Check inline comments

PR #670, opened 2022-07-02
Please review

YashTike:


jeelpatel18 (commented):
Good work Yash

jeelpatel18:
Good work Yash

NumberOfIslands.java
```diff
@@ -0,0 +1,52 @@
...
+entire grid. Whenever we reach an island i.e reach a 1, we mark the island as found and initiate DFS from that grid location
+to traverse the whole area of that island, till we reach the borders i.e reach 0.
+
+Note: The BFS solution is more optimal with space
+
+Did this code run on leetcode: Yes
+*/
+class Solution {
+    //Time Complexity: O(m*n)
+    //Space Complexity: O(m*n)
```
jeelpatel18, line 12: Why space is O(m*n). You are not creating any matrix over here right!!!!!

YashTike: Its the auxillary stack space. In the worst case where the matrix is only filled with 1s, it will traverse through all the 1s in a single DFS, so all the individual cells will be pushed onto the recursion stack.

For BFS though, it would be O(min(m,n)) because it would only check one level at a time.

jeelpatel18: Yes correct

PR #723, opened 2022-08-25
Completed DFS-2 - Please Review

adityaahale:


DecodeString.py
```diff
@@ -0,0 +1,33 @@
+# Time Complexity : O(N) Number of characters in string
```
jeelpatel18, line 1: Max(K's) * n

NumberOfIslands.py
```diff
@@ -0,0 +1,38 @@
+# Time Complexity : O(M * N) As it traverse the whole array
```
jeelpatel18, line 1: For dfs, in the worst case if it traverses the whole array then it would be m^2 * n^2

jeelpatel18 (commented):
Good work.

jeelpatel18:
Good work.

PR #731, opened 2022-09-17
decode string and number of islands completed. Please review.

shrutika-mahurkar:


daredevilteja (changes requested):
Please mention TC, SC and whether it ran successfully on leetcode or not, for every problem

daredevilteja:
Please do the requested changes

PR #919, opened 2023-10-04
Please review DFS-2

anaghayn:


Nandha2Tech (commented):
Great Work! For number of islands SC would be min between m and n.

Nandha2Tech:
Great Work! For number of islands SC would be min between m and n.

PR #969, opened 2024-01-24
Completed DFS-2

galibhaskar:
Solved:
200. Number of Islands
394. Decode String


Nandha2Tech (commented):
TC for Number of Islands problem would be O(2 m*n). Iteration of m*n will happen for two times in the worst case scenario.

Nandha2Tech:
TC for Number of Islands problem would be O(2 m*n). Iteration of m*n will happen for two times in the worst case scenario.

PR #1051, opened 2024-08-28
completed DFS-2

dhikki-study:
#########200. Number of Islands############################################################################
// Time Complexity : O(n*m)
// Space Complexity : O(1) -> as we are updating grid itself
// Did this code successfully run on Leetcode : Yes
// Any problem you faced while coding this : no problem

// Your code here along with comments explaining your approach in three sentences only
We will traverse through all element and we will encounter 1 we will apply DFS to check how many 1 are connected to it, also while searching for other 1 we will change current 1 to 0 this will help to ensure that once any island is counted it is not recounted and also we will keep a counter when ever 1 set of dis is complete and we look out for another 1 than counter is increased

########394. Decode String###################################################################################
// Time Complexity : O(n) -> as we are going through list just once
// Space Complexity : O(n) -> for 2 stacks
// Did this code successfully run on Leetcode : Yes
// Any problem you faced while coding this : no problem

// Your code here along with comments explaining your approach in three sentences only
Here we have maintained 2 stacks 1 stack for int and another for character, also we keep track of currentstring and current int which will help us to identify the number of string that we will put in stacks when and [ is encountered we insert the current value of currstring and current in stack and when  ] is encountered we poppet element for both stack and multiply by the times the int is present.

### DIFF
--- a/200. Number of Islands.py
+++ b/200. Number of Islands.py
@@ -1,0 +1,38 @@
+#########200. Number of Islands############################################################################
+// Time Complexity : O(n*m)
+// Space Complexity : O(1) -> as we are updating grid itself
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : no problem
+
+// Your code here along with comments explaining your approach in three sentences only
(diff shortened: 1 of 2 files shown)


Nandha2Tech (commented):
Great Work!

Nandha2Tech:
Great Work!

PR #1167, opened 2025-03-05
DFS-2 solutions in Java

DeviSuryaKumari:


Nandha2Tech (commented):
Great Work! Time complexity for DecodeString would be O(n * multiplication of all nested constants). Space complexity for NumberOfIslands problem with dfs would be O(m*n) which is recursive stack space.

Nandha2Tech:
Great Work! Time complexity for DecodeString would be O(n * multiplication of all nested constants). Space complexity for NumberOfIslands problem with dfs would be O(m*n) which is recursive stack space.

PR #1231, opened 2025-06-25
Completed DFS both problems

omkar04gaikwad:


super30admin:
Strengths:
- The student attempted to use DFS, which is a valid approach for this problem.
- The code structure is clear with separate functions for DFS and the main logic.

Areas for Improvement:
- The condition `r > 0 and c > n` in the DFS function is incorrect. It should check if `r` and `c` are within grid bounds (`0 <= r < m` and `0 <= c < n`).
- The DFS function should not always return `True`. It should perform the traversal without returning a boolean value.
- The direction handling is correct, but the overall logic needs to be fixed to properly mark visited cells and count islands.
- Consider using BFS instead of DFS for better space complexity in some cases, as shown in the reference solution.

PR #1263, opened 2025-09-25
Add solutions for Decode String and Number of Islands

ravithakur2k:
Added DecodeString.py implementing a stack-based approach to decode encoded strings, and NumberOfIslands.py with both BFS and DFS solutions to count islands in a grid. Both files include time and space complexity comments and problem-solving intuition.

### DIFF
--- a/DecodeString.py
+++ b/DecodeString.py
@@ -1,0 +1,36 @@
+# Time Complexity : O(n*k) where n is the length of the string and k is the max repeat count
+# Space Complexity : O(n*k)
+# Did this code successfully run on Leetcode : Yes
+# Any problem you faced while coding this : No
+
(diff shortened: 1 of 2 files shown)


super30admin:
- Strengths:
  - Correct implementation of both BFS and DFS approaches.
  - Clear and concise code with good comments.
  - Accurate time complexity analysis for BFS.
  
- Areas for Improvement:
  - The space complexity for DFS should be corrected to O(m × n) in the worst case due to recursion depth.
  - Consider mentioning the potential for stack overflow in DFS for very large grids and suggesting an iterative DFS as an alternative.
  - The problem title in the student's solution is "DFS-2", but the reference solution is for "Number of Islands". Ensure the problem title matches the actual problem being solved.

PR #1265, opened 2025-09-26
DFS-2

v-hemish:


super30admin:
- For problem1.py, you need to fix the condition in the BFS loop to check `grid[nx][ny] == '1'` instead of `grid[i][j] == '1'`. This is a critical bug that prevents the solution from working correctly.
- The rest of the implementation for problem1.py is good, with proper use of BFS and direction vectors.
- The decodeString solution in problem2.py is well-implemented, with clear logic and proper use of stacks. The variable names are descriptive, and the flow is easy to follow.
- Consider adding comments to explain the logic in both solutions, especially for more complex parts like the BFS loop or the stack operations in decodeString.

PR #1267, opened 2025-09-26
DFS-2 completed

sathvikareddy18:


super30admin:
- The main issue in your `numIslands` function is the incorrect indentation of the `while q:` loop. This causes the BFS to run for every cell, not just when a new island is found, leading to incorrect island counts.
- The rest of the implementation is clean and follows a good structure. The use of direction vectors and a queue for BFS is appropriate.
- Make sure to test your code with different grid configurations to catch such logical errors.
- The `decodeString` function is correct and well-implemented, showing good understanding of stack usage for this problem.

PR #1292, opened 2026-01-01
done

PurvaT-11:


super30admin:
- Problem1.py: Your solution is correct and well-structured. However, you can optimize the space complexity by modifying the grid in place to mark visited cells (e.g., changing '1' to '0' after visiting) instead of using a set. This would reduce the space complexity to O(min(m, n)), matching the reference solution.
- Problem2.py: Your solution is excellent. The logic is clear, and the implementation is efficient. Keep up the good work!

PR #1294, opened 2026-01-25
My_solutions

SrinivasaSupreeth:


super30admin:
Strengths:
- The base case in the DFS function is well-defined.

Areas for Improvement:
- The space complexity of the DFS recursion might be problematic for large grids. Consider using an iterative approach (like BFS or iterative DFS) to avoid potential stack overflow.
- Alternatively, you could use a stack for iterative DFS to maintain similar logic without recursion.
- The solution does not require the inner function to be defined inside numIslands. While it is acceptable in Python, it is worth noting that defining it outside might be slightly more efficient, but it is not a major issue.

Example of iterative DFS using a stack:

```
    if not grid:
        return 0

    rows, cols = len(grid), len(grid[0])
    count = 0
    stack = []

    for r in range(rows):
        for c in range(cols):
            if grid[r][c] == '1':
                count += 1
                stack.append((r, c))
                while stack:
                    i, j = stack.pop()
                    if i < 0 or i >= rows or j < 0 or j >= cols or grid[i][j] != '1':
                        continue
                    grid[i][j] = '0'
                    stack.append((i+1, j))
                    stack.append((i-1, j))
                    stack.append((i, j+1))
                    stack.append((i, j-1))

    return count
```

This iterative approach avoids recursion and uses a stack, which might be more efficient in terms of space.

PR #1305, opened 2026-02-09
Complete DFS-2

paridhimalviya:
Problem1 (https://leetcode.com/problems/number-of-islands/)

Problem2 (https://leetcode.com/problems/decode-string/)

### DIFF
--- a/DecodeString.swift
+++ b/DecodeString.swift
@@ -1,0 +1,120 @@
+//
+//  DecodeString.swift
+//  DSA-Practice
+//
+//  Created by Paridhi Malviya on 1/29/26.
+//
+
+class DecodeString {
+    private class StackUsingLL<T> {
+        private class LinkedListNode<T> {
+            var value: T
+            var next: LinkedListNode<T>?
+            init(value: T) {
+                self.value = value
+            }
+        }
+
+        private var head: LinkedListNode<T>?
+
+        func push(val: T) {
+            let newNode = LinkedListNode(value: val)
+            if (head == nil) {
+                //empty LinkedList
+                head = newNode
+                return
+            }
+
+            newNode.next = head
+            head = newNode
+        }
+
+        func pop() -> T? {
+            if (head == nil) {
+                //ll is empty
+                return nil
+            }
+            let value = head?.value
+            head = head?.next
+            return value
+        }
+    }
+    
+    /*
+     cna be done iteratively using stack.
+     can be done recursively using DFS
+     
+     String -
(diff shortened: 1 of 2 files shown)


super30admin:
Strengths:
- You have correctly implemented both BFS and DFS solutions for the problem.
- Your code is well-commented and easy to follow.
- You have considered the time and space complexities and provided explanations in comments.

Areas for Improvement:
- While your custom queue implementation is correct, in Swift, you can use an array and use `removeFirst()` for dequeuing. However, note that `removeFirst()` is O(n) operation, which would make the BFS less efficient. Alternatively, you can use a double-stack or a more efficient queue structure. But for this problem, the grid size is limited (300x300), so using an array might be acceptable. However, your linked list queue is efficient.
- For the DFS solution, the worst-case space complexity (O(m*n)) might lead to a stack overflow for large grids. It is generally safer to use BFS for this problem to avoid recursion depth issues.
- The DFS function uses recursion and might benefit from tail recursion optimization, but Swift does not guarantee tail recursion optimization. Consider using iterative DFS with a stack if you want to avoid recursion.
- In the BFS solution, you have defined the directions inside the while loop. It would be more efficient to define it outside the loop to avoid recreating it for every node.

super30admin:
Strengths:
- You have implemented both BFS and DFS solutions, showing a good understanding of different approaches.
- Your code is well-commented, and you have correctly analyzed the time and space complexities.
- The BFS solution uses a custom queue implementation, which demonstrates your understanding of data structures.

Areas for Improvement:
- Consider using Swift's standard data structures (like arrays) for queues and stacks to improve efficiency and reduce code complexity. For example, you can use an array as a queue by using `append` and `removeFirst`, but note that `removeFirst` is O(n). Alternatively, you can use two stacks or a linked list from a standard library if available. However, for this problem, the grid size is limited to 300x300, so even O(n) operations might be acceptable.
- The DFS solution uses recursion, which may lead to a stack overflow for very large grids. An iterative DFS using a stack would be more memory-efficient and avoid recursion depth issues.
- Separate unrelated problems (like DecodeString) into different files to maintain code organization and clarity.
- In the BFS solution, you dequeue an element and then iterate over directions. This is correct, but ensure that you mark the node as visited when you enqueue it (as you do) to avoid duplicate entries. This is correctly implemented.
- The custom queue implementation for BFS is correct, but it might be less efficient than using a standard array due to the overhead of linked list nodes. You can test with large grids to see if performance is acceptable.

PR #1308, opened 2026-02-20
solved decode string problem

jeetusingh247:


super30admin:
It appears that you have submitted a solution for the "Decode String" problem instead of the "Count of Islands" problem. Please ensure that you are solving the correct problem. For the "Count of Islands" problem, you need to implement a solution that traverses a 2D grid and counts connected components of land ('1') surrounded by water ('0'). 

A common approach is to use either BFS or DFS to mark visited lands to avoid double-counting. Here are some steps you might follow:
1. Iterate through each cell in the grid.
2. When you find a '1', increment the island count and then use BFS/DFS to mark all connected '1's as visited (e.g., by setting them to '0').
3. For BFS, you can use a queue to explore all adjacent cells (up, down, left, right). For DFS, you can use recursion or a stack.

Your current solution does not address this problem at all. Please review the problem statement and implement the correct solution.

PR #1310, opened 2026-03-07
DFS-2

shinjaneegupta:


super30admin:
- The solution for "NumberOfIslands" is correct in logic and efficient in terms of time and space complexity.
- However, you must import `deque` from `collections` to use it. Add: `from collections import deque` at the top.
- Also, ensure that you are only submitting the relevant code for the problem. The "Decode String" solution should not be included in the submission for "NumberOfIslands".
- The code structure is clear, but you should consider adding comments to explain the BFS process for better readability (though the code is already commented).
- Note that the problem expects the grid to be of characters (strings of length 1), and your solution correctly compares with '1' and sets to '0'. This is good.

PR #1311, opened 2026-03-07
DFS2 completed

sujaygijre:


super30admin:
Your solution for "numberOfIslands" is correct and efficiently handles the problem in terms of time complexity. However, there are a few points to consider for improvement:

- The DFS function returns a boolean, but you only use it to check if the starting cell was a land cell. Since you already check that `grid[i][j] == '1'` before calling `dfs`, the return value is redundant. You can change the DFS function to return void and remove the return statement. This would simplify the code.

- The recursive DFS might lead to stack overflow for very large grids (e.g., 300x300) because the recursion depth could be up to 90,000 in the worst case. While this might not happen in practice due to constraints, it's better to use an iterative approach (like BFS or iterative DFS) to avoid potential stack overflow. The reference solution uses BFS with a queue, which has a better worst-case space complexity.

- Consider using a separate visited array if modifying the input is not allowed. However, the problem allows modifying the input, so your approach is acceptable.

- In C++, using recursion for DFS is common, but for large grids, iterative methods are preferred. You could implement BFS using a queue or iterative DFS using a stack to improve space efficiency.

Overall, your solution is correct and clear, but for robustness, consider using an iterative method to handle large grids without recursion limits.

PR #1312, opened 2026-03-07
working solution

avcode3:


super30admin:
Your solution correctly solves the problem using a recursive DFS approach. Here are some strengths and areas for improvement:

Strengths:
- You correctly handle the base case for an empty grid.
- The algorithm efficiently marks connected components and counts islands.

Areas for Improvement:
- Space Complexity: The recursive DFS might lead to stack overflow for large grids (e.g., 300x300) because the recursion depth could be large. Consider using an iterative DFS or BFS to reduce space complexity. The reference solution uses BFS with a queue, which has a space complexity of O(min(m, n)) in the worst case.
- Function Naming: The function `checkNeighbour` is not very descriptive. Consider renaming it to something like `dfs` or `markIsland` to indicate its purpose.
- Parameter Passing: You are passing `row` and `col` to the recursive function, but these are constant for the entire grid. You could instead use the outer function's variables by making them attributes or using closure (though in Python, you can access them from the outer scope without passing if you are careful). Alternatively, you can define the recursive function inside `numIslands` and use nonlocal or capture the variables.
- Code Clarity: The recursive function modifies the grid in place, which is fine, but it might be clearer to have the function return void and only focus on marking.

Suggestion for Improvement:
You can implement an iterative BFS or DFS to avoid recursion. Here's an example of iterative BFS (similar to the reference solution) in Python:

```python
from collections import deque

class Solution(object):
    def numIslands(self, grid):
        if not grid:
            return 0
        rows = len(grid)
        cols = len(grid[0])
        count = 0
        directions = [(1,0), (-1,0), (0,1), (0,-1)]
        for i in range(rows):
            for j in range(cols):
                if grid[i][j] == '1':
                    count += 1
                    queue = deque([(i, j)])
                    grid[i][j] = '0'
                    while queue:
                        x, y = queue.popleft()
                        for dx, dy in directions:
                            nx, ny = x + dx, y + dy
                            if 0 <= nx < rows and 0 <= ny < cols and grid[nx][ny] == '1':
                                queue.append((nx, ny))
                                grid[nx][ny] = '0'
        return count
```

This iterative approach uses a queue and avoids recursion, thus reducing the space complexity.

PR #1313, opened 2026-03-08
Complete DFS-2

viveksharma180:


super30admin:
You have done a great job providing two different solutions (DFS and BFS) for the problem. Both are correct and well-implemented. Your comments on time and space complexity are accurate.

For the DFS solution, note that Python has a recursion limit (usually around 1000). For a grid of size 300x300, the recursion depth in the worst case could be 90000, which would exceed the recursion limit and cause a runtime error. Therefore, for this problem, the BFS solution is more practical and recommended.

Your BFS solution is efficient and follows the standard approach. It uses a queue to avoid recursion and has optimal space complexity.

One minor improvement: in the BFS solution, you are adding the starting cell to the queue and then immediately marking it as '0'. Then in the while loop, you process each cell. This is correct. However, you could also consider using a nested loop to avoid having to mark the starting cell outside the queue processing. But the current code is clear.

Overall, excellent work. Consider using BFS for problems with large grids to avoid recursion depth issues.

PR #1314, opened 2026-03-08
[Number of Islands && Decode String DFS]

takchiks:


super30admin:
- The student should consider using an iterative BFS or DFS (with a stack) to avoid potential stack overflow for large inputs.
   - Alternatively, if using DFS, note that the problem constraints (300x300) might be acceptable for recursion in some languages, but in Java the default stack size might not be enough. So it's better to use iterative methods.

PR #1315, opened 2026-03-08
Completed DFS-2

sarvanibaru:


super30admin:
Your solution for the "Count of Islands" problem has a good structure and follows the BFS approach correctly for the most part. However, there is a critical issue: you are not marking the starting cell as visited when you first encounter it. This means that during the BFS traversal, the starting cell remains '1', and when its neighbors are processed, they might add the starting cell back to the queue because it is still considered land. This can lead to duplicate processing and potentially infinite loops.

To fix this, you should mark the starting cell as visited (set it to '0') immediately when you add it to the queue. In your code, after `q.add(new int[]{i, j});`, you should add `grid[i][j] = '0';`. Alternatively, you can mark it when you pop it from the queue, but it is more efficient to mark it when you add it to avoid duplicate additions.

Here is the corrected version of your inner loop:
```java
if(grid[i][j] == '1') {
    count++;
    grid[i][j] = '0'; // Mark the starting cell as visited
    q.add(new int[]{i, j});
    while(!q.isEmpty()) {
        int[] curr = q.poll();
        for(int[] dir : dirs) {
            int newRow = dir[0] + curr[0];
            int newCol = dir[1] + curr[1];
            if(newRow >= 0 && newRow < m && newCol >= 0 && newCol < n && grid[newRow][newCol] == '1') {
                grid[newRow][newCol] = '0'; // Mark neighbor as visited when adding
                q.add(new int[]{newRow, newCol});
            }
        }
    }
}
```

Additionally, note that in the reference solution, the starting cell is marked as '0' immediately when found, and neighbors are marked when added to the queue. Your solution should do the same.

Your code for the decode string problem is correct

PR #1319, opened 2026-03-11
Problem1 added

megharaykar:


super30admin:
### Count of Islands (Problem1.py)
Your solution is correct and efficient. You have implemented the BFS approach correctly, and your code is clear and concise. Here are a few minor suggestions for improvement:

1. **Pythonic Queue Check**: Instead of `while len(q) > 0:`, you can use `while q:` which is more idiomatic in Python and equally efficient.
2. **Use Tuples for Directions**: The directions list is defined with tuples, which is good. However, when appending to the queue, you are using lists `[i,j]` and `[r,c]`. Consider using tuples for these as well, since they are immutable and slightly more efficient. For example, `q.append((i, j))` and then `curr = q.popleft()` would work the same, and `curr` would be a tuple. Alternatively, you can keep using lists if you prefer, but tuples are generally preferred for fixed-size data.
3. **Variable Naming**: The variable `dir` is a built-in function in Python. It is not a good practice to use it as a variable name. Consider renaming it to `d` or `direction` to avoid shadowing the built-in.

Overall, your solution is excellent. These are minor points and do not affect the correctness or performance significantly.

VERDICT: PASS

---

### Decoding String
It appears that you have submitted a solution for the "Number of Islands" problem instead of the "Decoding String" problem. Please double-check the problem statement and ensure you are solving the correct problem.

For the "Decoding String" problem, you need to implement a stack-based approach to handle nested brackets and repetitions. The reference solution provided uses two stacks: one for numbers and one for strings. Alternatively, you can use recursion with DFS.

Here are some key points for the correct solution:
- You should iterate through each character in the string.
- When you encounter a digit, you need to parse the entire number (as numbers can have multiple digits).
- When you encounter '[', push the current number and current string to stacks, then reset them.
- When you encounter ']', pop the number and string from the stacks, and append the current string repeated by the number to the popped string.
- For letters, simply append them to the current string.

Your current solution does not address any of these requirements. Please revisit the problem and implement the decoding logic.

VERDICT: NEEDS_IMPROVEMENT

PR #1324, opened 2026-05-10
Adding dfs-2 solutions

nikhylw:
solution 1 - W5_6_no_islands.py
solution 2 - W5_5_decode_string.py

### DIFF
--- a/W5_5_decode_string.py
+++ b/W5_5_decode_string.py
@@ -1,0 +1,37 @@
+class Solution:
+    def decodeString(self, s: str) -> str:
+
+        string_q = list()
+        number_q = list()
+        currstr = []
+        currnum = 0 
+
+        for char in s:
+
+            if char.isdigit():
+                currnum = currnum * 10 + int(char)
+            
+            elif char == "[":
+                number_q.append(currnum)
+                string_q.append(currstr)
+
+                currstr = []
+                currnum = 0                 
+
+            elif char == "]":
+                count = number_q.pop()
+                parent = string_q.pop()
+
+                for k in range(count):
+                    parent.extend(currstr)
+                
+                currstr = parent
+
+            else:
+                # Case: char is an alphabet
+                currstr.append(char)
+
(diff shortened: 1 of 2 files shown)


super30admin:
### Count of Islands (W5_6_no_islands.py)
**Strengths:**
- Correct DFS implementation that properly explores all four directions
- Good use of in-place grid modification to avoid visited tracking
- Clean separation between the main counting logic and the DFS helper
- Proper handling of boundary conditions

**Areas for Improvement:**
1. **Space complexity analysis:** Your claim of O(m * n) is technically correct for worst case, but be more precise. The recursion stack can go O(m * n) deep in worst case (entire grid is one island).
2. **Redundant parameters:** The `m` and `n` parameters in `dfs()` are unnecessary since you already store them as instance variables.
3. **Variable naming:** Avoid using `dir` as a variable name since it shadows Python's built-in. Use `direction` or `d` instead.
4. **Instance variable for count:** `self.count` could be a local variable in `numIslands()`, which would be cleaner and avoid potential thread-safety issues.

Overall, this is a solid solution that correctly solves the problem with good efficiency.

VERDICT: PASS

---

### Decoding String (W5_5_decode_string.py)
**Strengths:**
1. Clean, well-structured code with good variable naming
2. Correctly handles multi-digit numbers (e.g., "10", "100")
3. Proper use of stacks to maintain state at different nesting levels
4. Good handling of the nested case (e.g., "3[a2[c]]")
5. Clear comments explaining the approach

**Areas for Improvement:**
1. The time complexity comment is inaccurate - it should be O(n × k) where k is the maximum repeat count, not O(n + m)
2. Consider adding docstring to the method for better documentation
3. The variable name `string_q` could be more descriptive as `str_st` or `string_stack` to match the stack convention

**Minor Suggestions:**
- The solution is functionally equivalent to the reference solution and handles all edge cases correctly
- The use of Python lists as stacks is appropriate and idiomatic

VERDICT: PASS

